PR #15, opened 2020-12-09
CNV-8819 Enabling hotplug feature in Kubevirt

ydayagi:
Signed-off-by: Yaron Dayagi <ydayagi@redhat.com>

CNV-8819 As a user I want an easy method for enabling hotplug feature in Kubevirt

### DIFF
--- a/deploy/example/kubevirt-config.yaml
+++ b/deploy/example/kubevirt-config.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+data:
+  default-network-interface: masquerade
+  feature-gates: DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery,Sidecar,Snapshot,HotplugVolumes
+  selinuxLauncherType: virt_launcher.process
+  smbios: |-
+    Family: KubeVirt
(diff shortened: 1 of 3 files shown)


deploy/example/kubevirt-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+  feature-gates: DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery,Sidecar,Snapshot,HotplugVolumes
+  selinuxLauncherType: virt_launcher.process
+  smbios: |-
+    Family: KubeVirt
+    Manufacturer: KubeVirt
+    Product: None
+kind: ConfigMap
+metadata:
+  name: kubevirt-config
+  namespace: kubevirt
```
chenyosef, line 13: Please re-order the lines according to the convention (apiVersion, than kind, metadata and only then data)

chenyosef (changes requested):

deploy/example/kubevirt-config.yaml
ydayagi: Fixed

PR #21, opened 2021-01-11
Functional tests

rgolangh:
This PR add a functional test suite. Run it with `make test-functional`
The suite needs a working cluster and a `KUBECONFIG` env var, or passing --kubeconfig to the test execution.

Run using: `make test-functional`

The suite executes and tests:
- create a VM (centos at the moment)
- create a tenant cluster - installs vanilla upstream k8s in the VM
- deploy the CSI driver
- create storage class, PVC and a running pod that consumes a PV
- tears down


ydayagi (approved):

rgolangh:
Missing - using the CSI images under the test run

deploy/020-autorization.yaml
```diff
@@ -31,6 +31,9 @@ rules:
   - apiGroups: [""]
     resources: ["persistentvolumeclaims/status"]
     verbs: ["update", "patch"]
+  - apiGroups: [ "storage.k8s.io" ]
+    resources: [ "volumeattachments/status" ]
+    verbs: [ "get", "list", "watch", "update", "patch" ]
```
isaacdorfman, line 36: Are all the verbs necessary ?

deploy/020-autorization.yaml
```diff
@@ -105,17 +108,6 @@ rules:
...
----
-
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: openshift:csi-driver-controller-leader-election
-rules:
-  - apiGroups: [""]
-    resources: ["configmaps", "endpoints"]
-    verbs: ["get", "list", "watch", "update", "create", "delete"]
```
isaacdorfman, line 118: Are all the verbs necessary ?

tests/functional/functional_test.go
```diff
@@ -0,0 +1,339 @@
...
+				if p.Status.Phase == v1.PodRunning {
+					return true, nil
+				}
+			default:
+				return false, nil
+			}
+
+			return false, nil
+		},
+	)
```
isaacdorfman, line 264: I think that this part should be a separate function.

rgolangh: you didn't say which part exactly but I guess the until with sync?

hack/build-tests.sh
```diff
@@ -0,0 +1,26 @@
...
+export PATH=$PATH:$HOME/gopath/bin
+JOB_TYPE="${JOB_TYPE:-}"
+
+if [ "${JOB_TYPE}" == "travis" ]; then
+    go get -v -t ./...
+    go get github.com/mattn/goveralls
+    go get -v github.com/onsi/ginkgo/ginkgo
+    go get -v github.com/onsi/gomega
+    go get -u github.com/evanphx/json-patch
+    go mod vendor
```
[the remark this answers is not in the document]
rgolangh: Image has its own weight (creating downloding pushing downloading again) . Tests are anyway compiled, and they must be compatible with the code they test. If we had an image then we could easily enter inconsistency between the test and the code (code too new, dated tests)

tests/functional/functional_test.go
isaacdorfman: Yes

PR #24, opened 2021-01-21
OCPCNV-176 installer unique label

ydayagi:
[OCPCNV-176](https://issues.redhat.com/browse/OCPCNV-176) Resources created in infracluster should include "installer unique label"
In case tenant cluster which has PVCs created by CSI driver is destroyed by the installer, also the DVs related to those PVCs should be removed.

### DIFF
--- a/cmd/kubevirt-csi-driver/kubevirt-csi-driver.go
+++ b/cmd/kubevirt-csi-driver/kubevirt-csi-driver.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"strings"
 	"time"
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


cmd/kubevirt-csi-driver/kubevirt-csi-driver.go
```diff
@@ -23,6 +23,8 @@ var (
 	nodeName               = flag.String("node-name", "", "The node name - the node this pods runs on")
 	infraClusterNamespace  = flag.String("infra-cluster-namespace", "", "The infra-cluster namespace")
 	infraClusterKubeconfig = flag.String("infra-cluster-kubeconfig", "", "the infra-cluster kubeconfig file")
+	infraClusterLabelName  = flag.String("infra-cluster-labelname", "", "The infra-cluster label name to use when creating resources in infra cluster")
```
rgolangh, line 26: Two separate flags instead of one leaves room for error (set the key and not the value).
I think just a flag for a label is enough, even multi labels, just need validation 
For example `oc label` or `kubectl label` takes `KAY=VAL ... N_KEY=N_VAL` format

ydayagi: Currently we have just one label.
These are internal flags for the developer. Developer will test its usage and will not be able to miss the value.
However, if we do change it... do something like:
-label='labelA=valueA,labelB=valueB'
?

ydayagi: Done

PR #88, opened 2023-05-30
Make Controller Unpublish idempotent.

awels:
There was a race condition in unpublish that would sometimes get into an error state that it would never get out of causing unpublish to never happen even though the datavolume was unplugged from the VM already.

Update the k8s version used during testing to be more in line with the CSI version.
Fixes #

**Release note**:

```release-note
BugFix: Fixed issue with Controller Unpublish not being idem potent.
```


davidvossel (commented):
/lgtm
/approve

PR #90, opened 2023-07-06
Ensure volume is removed before returning success

awels:
When removing a volume from a VM, ensure that the volume is actually removed from the VM before returning nil from the ControllerUnpublishVolume function.
Added EnsureVolumeRemoved function to virtClient

Fixes #

**Release note**:

```release-note
NONE
```


davidvossel (approved):
/lgtm
/approve

PR #95, opened 2024-01-16
Use context when calling KV client 

nunnatsa:
The current code creates a TODO context instead of passing the context
from the request.

This PR changes the code to use the context from the request, so it will
support potential cancellation or timeout.

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/kubevirt/client.go
+++ b/pkg/kubevirt/client.go
@@ -21,15 +21,15 @@ type ClientBuilderFuncType func(kubeconfig string) (Client, error)
 // Client is a wrapper object for actual infra-cluster clients: kubernetes and the kubevirt
 type Client interface {
 	Ping(ctx context.Context) error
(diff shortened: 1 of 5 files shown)


awels (commented):
So CDI automatically applies the storage profile, why are we explicitly setting the value? If we leave it blank, then CDI will apply it from the storage profile. I am trying to understand what the purpose of this is.

nunnatsa:
> So CDI automatically applies the storage profile, why are we explicitly setting the value? If we leave it blank, then CDI will apply it from the storage profile. I am trying to understand what the purpose of this is.

Let me see if I understand: does CDI already sets the DV accessMode according to the storageClass?

awels (approved):

davidvossel (commented):
/approve
/lgtm

thanks!

PR #105, opened 2024-03-26
Get storage class from PVC for snapshot matching

awels:
Instead of using the PV to get the storage class use the PVC storage class instead. If it is nil then k8s version is < 1.28 and we return an error indicating we cannot determine the storage class name.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -141,9 +141,6 @@ kind: ClusterRole
 metadata:
   name: kubevirt-csi-snapshot
 rules:
-- apiGroups: [""]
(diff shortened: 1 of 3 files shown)


davidvossel (commented):
does the PV rbac need to be removed during the deploy steps for install now?

awels:
Good point removed.

davidvossel (commented):
/lgtm
/approve

davidvossel (commented):
/hold

PR #107, opened 2024-04-18
Allow RWX, FS infra PVC

nunnatsa:
For RWX pvc request, don't limit the infra PVC to be Block, but also allow FS.

**Release note**:

```release-note
For RWX pvc request, don't limit the infra PVC to be Block, but also allow FS.
```

### DIFF
--- a/e2e/create-pvc_test.go
+++ b/e2e/create-pvc_test.go
@@ -20,6 +20,7 @@ import (
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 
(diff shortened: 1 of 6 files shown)


e2e/create-pvc_test.go
```diff
@@ -173,6 +173,70 @@ var _ = Describe("CreatePVC", func() {
 		deletePod(tenantClient.CoreV1(), namespace, writerPod.Name)
 	})
 
+	It("should creates a RW-Many block pvc, with FS infra storage class, and attaches to pod", Label("pvcCreation", "RWX", "Block", "infra-FS"), func() {
+		pvcName := "test-pvc"
+		storageClassName := "infra-fs"
+		pvc := pvcSpec(pvcName, storageClassName, "10Mi")
+		volumeMode := k8sv1.PersistentVolumeBlock
+		pvc.Spec.VolumeMode = &volumeMode
```
awels, line 181: Nit: maybe use ptr.To here instead of a separate variable?

awels (commented):
looks good to me, just one little nit.

e2e/create-pvc_test.go
nunnatsa: Thanks @awels . good point - fixed

awels (approved):
/approve

awels:
Looks like maybe you need a rebase? We made some changes to the RBAC in particular for volume snapshots.

PR #108, opened 2024-04-18
Lazy load volume/snapshot mapping

awels:
Instead of eagerly calculating the mapping do it lazily when the mapping is requested. This will allow the driver to start when the RBAC in the tenant cluster might not be complete yet. Also added some debugging to allow one to determine why a mapping cannot be found easier.

Fixes #

**Release note**:

```release-note
Lazy load storage class/snapshot class mapping between infra and tenant clusters.
```


davidvossel (approved):
/lgtm
/approve

PR #124, opened 2025-01-14
Support Zone-Aware Topologies in the KubeVirt CSI Driver

moadqassem:
This pull request introduces support for zone-aware topology in the KubeVirt CSI driver. This enhancement allows the CSI driver to handle storage provisioning and attachment in multi-zone Kubernetes clusters. Integrates with Kubernetes topology-aware volume provisioners to ensure volumes are created in the same zone/region as the requesting node by:

- Implements support for TopologyKeys to enable zone/region-specific volume scheduling and attachment
- Modifies the Controller plugin to handle topology constraints during volume provisioning
- Updates to the Node plugin to include zone information in NodeGetInfo responses

Fixes #

**Release note**:

```release-note
Support zone and region-aware topologies in the KubeVirt CSI Driver
```

### DIFF
--- a/cmd/kubevirt-csi-driver/kubevirt-csi-driver.go
+++ b/cmd/kubevirt-csi-driver/kubevirt-csi-driver.go
@@ -105,7 +105,11 @@ func handle() {
 		klog.Fatal(err)
 	}
 
-	var nodeID string
+	var (
+		nodeID            string
+		allowedTopologies = map[string]string{}
+	)
+
 	if *nodeName != "" {
 		node, err := tenantClientSet.CoreV1().Nodes().Get(context.TODO(), *nodeName, v1.GetOptions{})
 		if err != nil {
@@ -121,6 +125,11 @@ func handle() {
 		}
 		nodeID = fmt.Sprintf("%s/%s", vmNamespace, vmName)
 		klog.Infof("Node name: %v, Node ID: %s", *nodeName, nodeID)
+		// systemUUID is the VM ID
+		nodeID = node.Status.NodeInfo.SystemUUID
+		allowedTopologies[service.WellKnownZoneTopologyKey] = node.Labels[service.WellKnownZoneTopologyKey]
+		allowedTopologies[service.WellKnownRegionTopologyKey] = node.Labels[service.WellKnownRegionTopologyKey]
+		klog.Infof("Node name: %v, Node ID: %s", nodeName, nodeID)
 	}
 
 	identityClientset = tenantClientSet
@@ -138,7 +147,8 @@ func handle() {
 		storageClassEnforcement,
 		nodeID,
 		*runNodeService,
-		*runControllerService)
+		*runControllerService,
+		allowedTopologies)
 
 	driver.Run(*endpoint)
 }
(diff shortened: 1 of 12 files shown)


moadqassem:
@awels  I have fixed some tests files that were failing and update the PR to resolve few conflicts. However, there were some other issues which I believe are not relevant to my PR:
```I0114 23:43:26.714262   37496 create-pvc_test.go:608] Event: AttachVolume.Attach failed for volume "tenant-pv" : rpc error: code = Unknown desc = invalid volume name [FailedAttachVolume]
```
And
```
 failed to provision volume with StorageClass "kubevirt": rpc error: code = InvalidArgument desc = non-block volume with RWX access mode is not supported
 ```

I also took a quick look into the testing process and In order to add an e2e k8s test to make sure that the zone and region are respected, I must change few things in the kubevirtci cluster provider(or add a new provider, haven't looked deep so not sure). This is where it needs to be adjusted to add labels that points the allowed topology: https://github.com/kubevirt/kubevirtci/blob/a291de27bb596074c79729ea6f88533555f523fd/cluster-up/cluster/ephemeral-provider-common.sh#L90

Let me know what do you think?

awels:
Sorry been very busy with other stuff, I will try to take a look at this soon.

awels:
If you look at the testing, we do actually exclude a few tests from the k8s test suite. In particular the RWX filesystem one, since we don't support that.

moadqassem:
> If you look at the testing, we do actually exclude a few tests from the k8s test suite. In particular the RWX filesystem one, since we don't support that.

I see. Alright let me check this filtering criteria.

awels:
https://github.com/kubevirt/csi-driver/blob/main/hack/run-k8s-e2e.sh#L117-L128 is what we use, I believe the last skip is what skips the RWX filesystem test.

awels:
Okay so I think I know what happened here, we made a change in how we identify which VM the volume is hotplugged into. Before we used the VMI, but that proved to be problematic, now we look at the VM directly. I am not sure if that affects how you look up VMs, it likely does.

moadqassem:
> Okay so I think I know what happened here, we made a change in how we identify which VM the volume is hotplugged into. Before we used the VMI, but that proved to be problematic, now we look at the VM directly. I am not sure if that affects how you look up VMs, it likely does.

hmm I don't think that I am listing or trying to get and VM as the code that I wrote is running on the DS which is in general deals with Node reference rather than the VM itself. The problem is, those tests are failing and I can't really check out what was the issue as you probably know the error traces(such as log messages) might come from other controllers which are not tracked. Also I noticed other PRs that have the exact failures. Do you know how we should proceed here ?

moadqassem:
After looking a deeper look, seems that these lines of code here broke the VMs lookups  to fetch VMs info from the CSINode object https://github.com/kubevirt/csi-driver/pull/130/files#diff-6cb93ab3a359a5bd44c54dbb81ac13c78f2051b868087906408564274b5e94bbR343-R349

moadqassem:
@awels can you please rerun the ci tests again. I have tested my recent changes and I was able to fix the problem,

awels:
PR looks good, one thing I forgot to ask about, did you enabled the topology tests in the k8s test suite. You can modify https://github.com/kubevirt/csi-driver/blob/main/hack/test-driver.yaml and set `topology: true` and that should be all that is needed to run the topology tests.

awels:
I created a branch here https://github.com/awels/kubevirt-csi/tree/topology_support with the changes I think are needed to complete this PR. I will also update one of the lanes to use the WFFC binding mode environment variable. If you could take the last commit on that branch and move it into your PR I would appreciate it. If you can't I will make a PR out of that branch and give you credit.

oujonny:
/reopen

@moadqassem can we get this finally merged

PR #125, opened 2025-01-16
Fix functional test that checks if tenant PVC cannot access unlabeled infra PVCs

awels:
This test creates a PVC in the infra cluster manually and then tries to access it from a tenant cluster by crafting a special PVC that references the infra PVC.

Since the labels will not match the PVC should not be accessible. After wards one will not be able to delete the tenant PV because there is no matching PVC in the infra cluster.

This removes the check that the tenant PV can be deleted.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2e/create-pvc_test.go
+++ b/e2e/create-pvc_test.go
@@ -110,7 +110,7 @@ var _ = Describe("CreatePVC", func() {
 		Entry("Block volume mode", Label("Block"), k8sv1.PersistentVolumeBlock, withBlock, blockAttachCommand),
 	)
 
-	It("should creates a RW-Many block pvc and attaches to pod", Label("pvcCreation", "RWX", "Block"), func() {
+	It("should create a RW-Many block pvc and attaches to pod", Label("pvcCreation", "RWX", "Block"), func() {
 		pvcName := "test-pvc"
 		storageClassName := "kubevirt"
 		pvc := pvcSpec(pvcName, storageClassName, "10Mi")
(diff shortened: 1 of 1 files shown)


akalenyu:
> Since the labels will not match the PVC should not be accessible

This was the key part for me.
That CVE test is negative, so the pod that mounts this volume is going to fail (and that is asserted).

What still feels off to me is that I don't understand why this started failing just now.
Unless, some of the commits went in without CI? CSI Sidecar bumps that contain changes in behavior?
Maybe this commit? https://github.com/kubevirt/csi-driver/commit/906c86e7f53f30a0ea29b66575ad4ce8b7d2ed48

Sorry for being a pain I am hesitant to give out an lgtm here without having rock solid understanding of what it is that made it start failing now when it should have in fact failed when the test was introduced

awels:
Yeah I honestly don't know either, that commit did add the code so it is likely that is culprit. But it passed CI without issues.

akalenyu:
> Yeah I honestly don't know either, that commit did add the code so it is likely that is culprit. But it passed CI without issues.

How come the PV is even around if we're using `PersistentVolumeReclaimDelete`? https://github.com/kubevirt/csi-driver/blob/35836e0c8b68d9916d29a838ea60cdd3fc6199cf/e2e/create-pvc_test.go#L572

akalenyu:
Can https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/2644-honor-pv-reclaim-policy be related somehow? It went beta in 1.31.

awels:
I don't think so, the problem is that when we delete the PV in the tenant cluster, it can't find (correctly) the matching PV in the infra cluster, and returns an error. This causes the PV to not get deleted.

akalenyu (approved):

PR #129, opened 2025-03-04
Use proper mount util library k8s.io/utils/mount->k8s.io/mount-utils

akalenyu:
We should be using the new location of this library
since the old one is not receiving updates for a while now https://github.com/kubevirt/csi-driver/blob/ae297e73c0ee982d80fd8d677db66bc2fa370342/vendor/k8s.io/utils/mount/README.md?plain=1#L4-L9

Fixes #

**Release note**:

```release-note
Use correct mount-utils k8s.io/utils/mount->k8s.io/mount-utils
```


awels (approved):
/approve

PR #130, opened 2025-03-06
Ensure volume stays attached through reboots

akalenyu:
While this is not stated explicitly in the CSI spec,
k8s expects ControllerPublished volumes to stay attached regardless of reboots.

Today, we lose the attachment if a reboot is quick enough that a failover to a different node was not initiated
(default node.kubernetes.io/unreachable toleration is 5 minutes)
and there is no extra invocation of the publish, resulting in an eternal failure.

We can align to this implicit guideline by hotplugging on the VM level:
https://kubevirt.io/user-guide/storage/hotplug_volumes/#retain-hotplugged-volumes-after-restart

Fixes #
https://issues.redhat.com/browse/OCPBUGS-50562

**Release note**:

```release-note
BugFix: Ensure volume stays attached through reboots
```

### DIFF
--- a/pkg/kubevirt/client.go
+++ b/pkg/kubevirt/client.go
@@ -55,6 +55,7 @@ type Client interface {
 	Ping(ctx context.Context) error
 	ListVirtualMachines(ctx context.Context, namespace string) ([]kubevirtv1.VirtualMachineInstance, error)
 	GetVirtualMachine(ctx context.Context, namespace, name string) (*kubevirtv1.VirtualMachineInstance, error)
+	GetWorkloadManagingVirtualMachine(ctx context.Context, namespace, name string) (*kubevirtv1.VirtualMachine, error)
 	DeleteDataVolume(ctx context.Context, namespace string, name string) error
 	CreateDataVolume(ctx context.Context, namespace string, dataVolume *cdiv1.DataVolume) (*cdiv1.DataVolume, error)
 	GetDataVolume(ctx context.Context, namespace string, name string) (*cdiv1.DataVolume, error)
@@ -200,10 +201,10 @@ func (c *client) EnsureVolumeAvailable(ctx context.Context, namespace, vmName, v
 		}
 		for _, volume := range vmi.Status.VolumeStatus {
 			if volume.Name == volumeName && volume.Phase == kubevirtv1.VolumeReady {
-				return true, nil
+				return ensureVolumeAvailableVM(ctx, c, namespace, vmName, volumeName)
 			}
 		}
-		// Have not found the ready hotplugged volume
+
 		return false, nil
 	})
 }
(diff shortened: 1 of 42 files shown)


pkg/service/controller.go
```diff
@@ -480,13 +498,19 @@ func (c *ControllerService) ControllerUnpublishVolume(ctx context.Context, req *
...
 
 func (c *ControllerService) removeVolumeFromVm(ctx context.Context, dvName, vmName string) error {
-	vm, err := c.virtClient.GetVirtualMachine(ctx, c.infraClusterNamespace, vmName)
+	vm, err := c.virtClient.GetWorkloadManagingVirtualMachine(ctx, c.infraClusterNamespace, vmName)
 	if err != nil {
-		return err
+		if !errors.IsNotFound(err) {
+			return err
+		}
+		return status.Errorf(codes.NotFound, "vm %s not found", vmName)
```
awels, line 506: log and return nil, nil instead?

pkg/service/controller_test.go
```diff
@@ -790,13 +791,7 @@ func (c *ControllerClientMock) ListVirtualMachines(_ context.Context, namespace
...
 				Namespace: namespace,
 			},
-			Spec: kubevirtv1.VirtualMachineInstanceSpec{
-				Domain: kubevirtv1.DomainSpec{
-					Firmware: &kubevirtv1.Firmware{
-						UUID: types.UID(testVMUID),
-					},
-				},
-			},
+			Spec: kubevirtv1.VirtualMachineInstanceSpec{},
```
awels, line 794: Do we even care about the Spec at this point? Just don't define it?

pkg/service/controller.go
```diff
@@ -396,15 +402,22 @@ func (c *ControllerService) ControllerPublishVolume(
...
 
 func (c *ControllerService) isVolumeAttached(ctx context.Context, dvName, vmName string) (bool, error) {
-	vm, err := c.virtClient.GetVirtualMachine(ctx, c.infraClusterNamespace, vmName)
+	vm, err := c.virtClient.GetWorkloadManagingVirtualMachine(ctx, c.infraClusterNamespace, vmName)
 	if err != nil {
-		return false, err
+		if !errors.IsNotFound(err) {
+			return false, err
+		}
+		return false, status.Errorf(codes.NotFound, "vm %s not found", vmName)
```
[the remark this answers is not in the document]
akalenyu: if the VM is not found returning `false, nil` will result in a hotplug attempt to a non existing VM

pkg/kubevirt/client.go
```diff
@@ -486,5 +502,41 @@ func appendVolumeSnapshotInfraTenantMapping(mapping *InfraTenantStorageSnapshotM
 	return mapping
 }
 
+func ensureVolumeRemovedVM(ctx context.Context, c *client, namespace, name, volumeName string) (bool, error) {
+	vm, err := c.virtClient.KubevirtV1().VirtualMachines(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
```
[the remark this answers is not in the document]
akalenyu: yup

pkg/service/controller.go
akalenyu: yep

pkg/service/controller_test.go
akalenyu: right

pkg/service/controller.go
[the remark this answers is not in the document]
awels: Okay

awels (approved):
/approve

PR #131, opened 2025-03-20
Resize for extra capacity when restoring to greater target volume

akalenyu:
While we do allow restoring to a greater target volume,
this was a no-op for fs volumes, because we never resized the fs
to make use of the new capacity.

This change always resizes the fs if necessary on volume publish.

Fixes #

**Release note**:

```release-note
BugFix: resize filesystems to make use of volumes that were restored to bigger target
```

### DIFF
--- a/pkg/service/node.go
+++ b/pkg/service/node.go
@@ -12,6 +12,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	mount "k8s.io/mount-utils"
+	utilexec "k8s.io/utils/exec"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 
@@ -30,6 +31,7 @@ type NodeService struct {
 	deviceLister DeviceLister
 	fsMaker      FsMaker
 	mounter      mount.Interface
+	resizer      ResizerInterface
 	dirMaker     dirMaker
 }
 
@@ -41,8 +43,17 @@ type dirMaker interface {
 	Make(path string, perm os.FileMode) error
 }
 
+type ResizerInterface interface {
+	NeedResize(devicePath string, deviceMountPath string) (bool, error)
+	Resize(devicePath, deviceMountPath string) (bool, error)
+}
+
 var NewMounter = func() mount.Interface {
 	return mount.New("")
+}
+
+var NewResizer = func() ResizerInterface {
+	return mount.NewResizeFs(utilexec.New())
 }
 
 var NewDeviceLister = func() DeviceLister {
@@ -65,6 +76,7 @@ func NewNodeService(nodeId string) *NodeService {
 		deviceLister: NewDeviceLister(),
 		fsMaker:      NewFsMaker(),
 		mounter:      NewMounter(),
+		resizer:      NewResizer(),
 		dirMaker: dirMakerFunc(func(path string, perm os.FileMode) error {
 			// MkdirAll returns nil if path already exists
 			return os.MkdirAll(path, perm)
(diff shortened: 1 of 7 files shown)


sanity/sanity_test.go
```diff
@@ -354,3 +354,21 @@ type fakeFsMaker struct{}
...
+
+func (r *fakeResizer) Resize(devicePath, deviceMountPath string) (bool, error) {
+	if r.resizedMap == nil {
+		r.resizedMap = map[string]struct{}{}
+	}
+	key := fmt.Sprintf("%s:%s", devicePath, deviceMountPath)
+	r.resizedMap[key] = struct{}{}
+
+	return true, nil
+}
```
akalenyu, line 370: @awels not completely sure if the sanity suite does something with this? I saw a similar impl. in the `Mount()` call

pkg/service/node.go
```diff
@@ -114,31 +126,34 @@ func (n *NodeService) NodeStageVolume(_ context.Context, req *csi.NodeStageVolum
...
-	if req.VolumeCapability.GetMount() != nil {
-		// Filesystem volume mode, create FS if needed
-		// get the VMI volumes which are under VMI.spec.volumes
-		// serialID = kubevirt's DataVolume.UID
-		device, err := getDeviceBySerialID(req.VolumeContext[serialParameter], n.deviceLister)
-		if err != nil {
-			klog.Errorf("Failed to fetch device by serialID %s", req.VolumeId)
-			return nil, err
-		}
+	if req.VolumeCapability.GetMount() == nil {
```
akalenyu, line 129: sorry for this mess, forgot I change this. It's just flipping the logic to return early for better indentation
using `if req.VolumeCapability.GetMount() == nil {`. no logical change

pkg/service/node.go
```diff
@@ -236,34 +251,67 @@ func (n *NodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	if isBlock {
+		err := n.ensureMountFileExists(targetPath)
 		if err != nil {
-			return nil, err
+			return err
 		}
 	} else {
 		// MkdirAll returns nil if path already exists
-		err = n.dirMaker.Make(targetPath, 0750)
+		err := n.dirMaker.Make(targetPath, 0750)
```
awels, line 299: Can you make this
```go
if err := n.dirMaker.Make(targetPath, 0750); err != nil {
  return err
}

File: pkg/service/node.go [hunk 4]
```diff
@@ -236,34 +251,67 @@ func (n *NodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
-			return nil, err
+			return err
 		}
 		klog.V(3).Infof("GetMount() %v", req.VolumeCapability.GetMount())
 		klog.V(3).Infof("Mounting devicePath %s, on targetPath: %s with FS type: %s",
 			device, targetPath, fsType)
 	}
 
-	err = n.mounter.Mount(device.Path, targetPath, fsType, mountOptions)
+	err := n.mounter.Mount(device.Path, targetPath, fsType, mountOptions)
```
awels, line 308: Same just scope the err variable to the if statement.

sanity/sanity_test.go
awels: For the mount call if it calls `Unmount` or `List`, then saving the value in the `Mount` call makes sense so we have something to return. For resize it either succeeds or not. And the sanity suite is only interested in the successful calls I believe, so no real need to save the keys of the paths we are 'resizing'

awels (commented):
looks good, a few nitpicks, but nothing that makes me want to reject the PR.

akalenyu:
> @akalenyu: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-csi-driver-split-k8s-suite-k8s	[e538c09](https://github.com/kubevirt/csi-driver/commit/e538c09900af512c59905dce46d83e8bcdff74b5)	[link](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_csi-driver/131/pull-csi-driver-split-k8s-suite-k8s/1902783380844449792)	true	`/test pull-csi-driver-split-k8s-suite-k8s`

I think the first invocation of the test binary (test-driver.yaml) sometimes accidentally whacks our RBAC, breaking the second invocation (test-rwx.yaml)

pkg/service/node.go
akalenyu: done

pkg/service/node.go
akalenyu: yup

sanity/sanity_test.go
akalenyu: ah okay gotcha

awels: No harm done if we do save it, I don't think it will ever be used for anything.

awels (approved):
/approve

PR #134, opened 2025-04-06
Ensure unpublish works for volumes carried over upgrades

akalenyu:
One upgrade consideration missed in PR 130 is that
ControllerUnpublish could be called against a VMI-level-hotplugged
carry over from an older driver version.
In that case, we would not issue a removevolume call
at all and fail in ensuring the volume got removed.

ControllerPublish doesn't suffer from a similar issue
since the volume would just get added on the VM level in addition.

Fixes #

**Release note**:

```release-note
Follow up vm level hotplug by ensuring correct handling of carry over from older versions
```


awels (approved):
/lgtm
/approve

PR #138, opened 2025-04-20
Implement volume expansion via infra PVC resize

### LINKED ISSUE #99: kubevirt sc support ALLOWVOLUMEEXPANSION?
**Is your feature request related to a problem? Please describe**:
A clear and concise description of what the problem is.
When this infra cluster sc is ceph rbd, it support volume expansion, so if we use kubevirt csi driver，we hope can also use this feature.
**Describe the solution you'd like**:
kubevirt sc support ALLOWVOLUMEEXPANSION.
**Describe alternatives you've considered**:

**Additional context**:

akalenyu:
This commit implements volume expansion via resizing
of the underlying PVC on the infra cluster.

Fixes #99 

**Release note**:

```release-note
kubevirt-csi volume expansion support
```

### DIFF
--- a/deploy/controller-infra/base/deploy.yaml
+++ b/deploy/controller-infra/base/deploy.yaml
@@ -145,6 +145,27 @@ spec:
             requests:
               memory: 20Mi
               cpu: 10m
+        - name: csi-resizer
+          image: registry.k8s.io/sig-storage/csi-resizer:v1.13.1
+          args:
+            - "-csi-address=/csi/csi.sock"
+            - "-kubeconfig=/var/run/secrets/tenantcluster/value"
+            - "-v=5"
+            - "-timeout=3m"
+            - '-handle-volume-inuse-error=false'
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+            - name: tenantcluster
+              mountPath: "/var/run/secrets/tenantcluster"
+          resources:
+            requests:
+              cpu: 10m
+              memory: 20Mi
+          securityContext:
+            capabilities:
+              drop:
+                - ALL
       volumes:
         - name: socket-dir
           emptyDir: {}
--- a/deploy/controller-tenant/base/deploy.yaml
+++ b/deploy/controller-tenant/base/deploy.yaml
@@ -140,6 +140,24 @@ spec:
             requests:
               memory: 20Mi
               cpu: 10m
+        - name: csi-resizer
+          image: registry.k8s.io/sig-storage/csi-resizer:v1.13.1
+          args:
+            - "-csi-address=/csi/csi.sock"
+            - "-v=5"
+            - "-timeout=3m"
+            - '-handle-volume-inuse-error=false'
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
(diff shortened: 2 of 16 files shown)


deploy/controller-infra/base/deploy.yaml
```diff
@@ -145,6 +145,29 @@ spec:
...
+            - "-timeout=3m"
+            - '-handle-volume-inuse-error=false'
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+            - name: tenantcluster
+              mountPath: "/var/run/secrets/tenantcluster"
+          resources:
+            limits:
+              memory: 400Mi
```
awels, line 163: Any reason for this limit here, in general I know we want limits, but since this thing is watching a large number of resource in the cluster it can potentially grow large in large clusters. That is why for control plane pods we usually just set the requests not the limits.

deploy/controller-tenant/base/deploy.yaml
```diff
@@ -140,6 +140,26 @@ spec:
...
+            - "-csi-address=/csi/csi.sock"
+            - "-v=5"
+            - "-timeout=3m"
+            - '-handle-volume-inuse-error=false'
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+          resources:
+            limits:
+              memory: 400Mi
```
awels, line 155: Same as above

pkg/kubevirt/client.go
```diff
@@ -318,6 +347,43 @@ func (c *client) GetDataVolume(ctx context.Context, namespace string, name strin
...
+
+func (c *client) ExpandPersistentVolumeClaim(ctx context.Context, namespace string, claimName string, desiredSize int64) error {
+	currentPVC, err := c.GetPersistentVolumeClaim(ctx, namespace, claimName)
+	if err != nil {
+		return err
+	}
+	desiredQuantity := *resource.NewQuantity(desiredSize, resource.DecimalSI)
+	currentQuantity := currentPVC.Spec.Resources.Requests.Storage()
+	if currentQuantity.Cmp(desiredQuantity) >= 0 {
+		if currentQuantity != nil {
```
awels, line 372: This indicates that currentQuantity can be nil, but we just did a cmp on a potentially nil value? Maybe rewrite it like this?
```go
if currentQuantity != nil && currentQuantity.Cmp(desiredQuantity) >= 0 {
  klog.V(5).Infof("Volume %s of quantity %v is larger than requested quantity %v, no need to expand", claimName, *currentQuantity, desiredQuantity)
  return nil
} else if currentQuantity == nil {
  return nil
}
```

awels (commented):
Looks good a few minor comments nothing major.

deploy/controller-infra/base/deploy.yaml
akalenyu: No reason, fair point, removed

pkg/kubevirt/client.go
akalenyu: My bad, that nil check was redundant

awels (approved):
/approve

PR #143, opened 2025-06-30
Fix deadlock while reattaching volume

kvaps:

This PR adds logic that prevents the controller from trying to attach or detach a volume when the VMI status already shows the volume is in the desired state.

**Which issue(s) this PR fixes** 

We ran into a problem while a volume was being attached. The attacher logged:

```
I0630 08:45:50.046983       1 csi_handler.go:243] "Error processing" VolumeAttachment="csi-6da36ea7aa56330cd69c4061f0a7e0266a7e08d3dd819943acab55922f5e764d" err="failed to detach: rpc error: code = Internal desc = ControllerUnpublishVolume failed for pvc-0645b5aa-75c0-4f00-ac2d-a9998256fa65: Message: 'Resource 'pvc-0645b5aa-75c0-4f00-ac2d-a9998256fa65' is still in use.'; Cause: 'Resource is mounted/in use.'; Details: 'Node: plo-csxhk-004, Resource: pvc-0645b5aa-75c0-4f00-ac2d-a9998256fa65'; Correction: 'Un-mount resource 'pvc-0645b5aa-75c0-4f00-ac2d-a9998256fa65' on the node 'plo-csxhk-004'.'; Reports: '[685C0CB4-00000-002093]'"
```

On the node the disk was already attached to the VM, yet the corresponding VolumeAttachment had a deletion timestamp.

So we got stuck: the hp-volume pod couldn’t start because it was waiting for the VolumeAttachment to disappear, and the VolumeAttachment couldn’t be removed because the VM was still using the disk.

This may be a race: kubevirt-csi-controller thought the disk had been detached (the hp-volume pod was gone), while the VM was still holding it. Kubernetes then created a new VolumeAttachment in the tenant cluster, triggering another attach operation.

**Release note**:

```release-note
Fix deadlock while reattaching volume
```

### DIFF
--- a/pkg/kubevirt/client.go
+++ b/pkg/kubevirt/client.go
@@ -67,7 +67,9 @@ type Client interface {
 	RemoveVolumeFromVM(ctx context.Context, namespace string, vmName string, hotPlugRequest *kubevirtv1.RemoveVolumeOptions) error
 	RemoveVolumeFromVMI(ctx context.Context, namespace string, vmName string, hotPlugRequest *kubevirtv1.RemoveVolumeOptions) error
 	EnsureVolumeAvailable(ctx context.Context, namespace, vmName, volumeName string, timeout time.Duration) error
+	EnsureVolumeAvailableVM(ctx context.Context, namespace, name, volumeName string) (bool, error)
 	EnsureVolumeRemoved(ctx context.Context, namespace, vmName, volumeName string, timeout time.Duration) error
+	EnsureVolumeRemovedVM(ctx context.Context, namespace, name, volumeName string) (bool, error)
 	EnsureSnapshotReady(ctx context.Context, namespace, name string, timeout time.Duration) error
 	EnsureControllerResize(ctx context.Context, namespace, claimName string, timeout time.Duration) error
 	CreateVolumeSnapshot(ctx context.Context, namespace, name, claimName, snapshotClassName string) (*snapshotv1.VolumeSnapshot, error)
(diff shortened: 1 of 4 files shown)


awels (approved):
/approve

alifelan:
Was there anything that made you change your mind? Have we seen any issues related to this behaviour? While I don't have experience deploying a KubeVirt CSI Driver with these changes, at a first glance I believe we will be impacted by them. We run into timeouts frequently, and after the second attempt, our `VolumeAttachments` won't be reflecting the actual state

akalenyu:
> Was there anything that made you change your mind? Have we seen any issues related to this behaviour? While I don't have experience deploying a KubeVirt CSI Driver with these changes, at a first glance I believe we will be impacted by them. We run into timeouts frequently, and after the second attempt, our `VolumeAttachments` won't be reflecting the actual state

oh, I didn't change my mind - I never pulled this commit into the productized fork.
I do believe we want to take a step back with this change and discuss the problem/solutions in one of the sig-storage meetings. I agree with your assessment.

kvaps:
We have been using this patch in production for several months, and it works as expected. No more deadlocks — previously we consistently had problems with reattaching PVCs.

The original issue is described here: https://github.com/kubevirt/csi-driver/pull/143#issue-3189720038

> using different sources of truth (vm.Spec.Template.Spec.Volumes and vmi.Status.VolumeStatus)

We can be sure that a volume is already detached if it is no longer present in vmi.Status.VolumeStatus, and therefore consider the detach operation completed.
Likewise, we can be sure that a volume is already attached if it is present in vmi.Status.VolumeStatus, and in that case we do not attempt to attach it again.

If I recall correctly, the main point of this PR is introducing this logic. Without it, the volume may be detached by the driver, but remain stuck forever in case of operation fails by 2 minutes timeout.

PR #163, opened 2025-10-24
fix: make NodeUnpublishVolume idempotent

jaskiratsingh98:
Make NodeUnpublishVolume idempotent by only removing the mount if one exists.

The goal is to make NodeUnpublishVolume idempotent as required by the CSI spec. Previously, if a mount was already removed, this call would fail and would ultimately be re-queued. This would, for example, lead to pods getting stuck in a Terminating state."

Fixes #

**Release note**:

```release-note
NodeUnpublishVolume is now idempotent by ensuring a mount exists before trying to unmount.
```

### DIFF
--- a/pkg/service/node.go
+++ b/pkg/service/node.go
@@ -390,20 +390,45 @@ func (n *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	}
 
 	targetPath := req.GetTargetPath()
-	klog.V(5).Infof("Unmounting %s", targetPath)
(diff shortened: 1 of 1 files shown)


akalenyu (commented):
I was just tackling this too in #165, but apparently it is solved by using the proper mount utils library - https://github.com/kubevirt/csi-driver/blob/7cd7c19c32406833d6f1da93d0cbf259c1b1b1ec/vendor/k8s.io/mount-utils/mount_linux.go#L930-L933 (on most distros)

In which version are you seeing this? we could still ofc do better with the cleanup API

PR #166, opened 2025-11-17
Add DataVolume Finalizer

kellydx:
Add a finalizer to the DataVolume to prevent it from being accidentally deleted manually or by external workloads.

I encountered a scenario where some DataVolumes were accidentally deleted, leading to data loss and VMs being unable to boot. As a precaution, a DataVolume created by KVCSI should have a finalizer to prevent it from being accidentally deleted, ensuring resources are cleaned up properly.

Fixes #

**Release note**:

```
Added finalizer "cdi.kubevirt.io/user-volume-protection" to DataVolume.

```

### DIFF
--- a/pkg/kubevirt/client.go
+++ b/pkg/kubevirt/client.go
@@ -35,6 +35,7 @@ const (
 	annDefaultSnapshotClass         = "snapshot.storage.kubernetes.io/is-default-class"
 	InfraStorageClassNameParameter  = "infraStorageClassName"
 	InfraSnapshotClassNameParameter = "infraSnapshotClassName"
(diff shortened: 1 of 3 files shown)


mhenriks:
/lgtm
Thanks @kellydx!

awels (commented):
Looks like in some cases the finalizer is not removed from the infra PVC, and this is causing a ton of tests to fail.
You should be able to reproduce the issue with doing
```bash
make cluster-up
make cluster-sync-split
./hack/run-e2e.sh
```
This will run some of the tests and you should be able to see what is failing

PR #169, opened 2025-12-18
Add udev rescan on NodeStage

### LINKED ISSUE #152: Retrigger udev on failure to get device serial
**What happened**:
We implicitly rely on udev to gather the serial ID of devices as they are added to the guest. If udev fails, for example due to transient networking issues on the underlying data path, it does not retry. We are therefore stuck with failed mounts

NodeStageVolume can retrigger udev if it cannot find the serial ID, allowing us to eventually succeed once the underlying networking issues are resolved.

You can reproduce this in a slightly contrived way by adding a udev rule that will force udev to timeout, simulating a command timing out due to a networking issue. I did this by adding the following line to `/usr/lib/udev/rules.d/60-persistent-storage.rules`

```
KERNEL=="sd*[!0-9]|sr*", ENV{ID_SERIAL}!="?*", IMPORT{program}="/usr/bin/sleep 600"
```

After creating a pod and pvc on this tenant node, you can see mount errors in the pod events:
```
Warning  FailedMount             10s (x6 over 27s)  kubelet                  MountVolume.MountDevice failed for volume "pvc-0966c060-98fc-4037-93f9-3833b5874e98" : rpc error: code = Unknown desc = couldn't find device by serial id
```

and udev logs show that it quits:
```
Oct 03 17:01:20 vm-018c7ff6 systemd-udevd[2674348]: seq 7638 '/devices/pci0000:00/0000:00:02.4/0000:05:00.0/virtio1/host0/target0:0:0/0:0:0:0/block/sda' is taking a long time
Oct 03 17:01:20 vm-018c7ff6 systemd-udevd[2674348]: seq 7640 '/devices/pci0000:00/0000:00:02.4/0000:05:00.0/virtio1/host0/target0:0:0/0:0:0:3/block/sdd' is taking a long time
Oct 03 17:03:20 vm-018c7ff6 systemd-udevd[2674348]: seq 7638 '/devices/pci0000:00/0000:00:02.4/0000:05:00.0/virtio1/host0/target0:0:0/0:0:0:0/block/sda' killed
Oct 03 17:03:20 vm-018c7ff6 systemd-udevd[2674348]: seq 7640 '/devices/pci0000:00/0000:00:02.4/0000:05:00.0/virtio1/host0/target0:0:0/0:0:0:3/block/sdd' killed
Oct 03 17:03:20 vm-018c7ff6 systemd-udevd[2674348]: worker [2943916] terminated by signal 9 (KILL)
Oct 03 17:03:20 vm-018c7ff6 systemd-udevd[2674348]: worker [2943916] failed while handling '/devices/pci0000:00/0000:00:02.4/0000:05:00.0/virtio1/host0/target0:0:0/0:0:0:0/block/sda'
Oct 03 17:03:20 vm-018c7ff6 systemd-udevd[2674348]: worker [2943915] terminated by signal 9 (KILL)
Oct 03 17:03:20 vm-018c7ff6 systemd-udevd[2674348]: worker [2943915] failed while handling '/devices/pci0000:00/0000:00:02.4/0000:05:00.0/virtio1/host0/target0:0:0/0:0:0:3/block/sdd
```

Checking back periodically I see that it does not retry, as expected since it is event driven.

**Environment**:
Looking at the code I think this should happen in most envs and versions

zworman-goog:
When KubeVirt adds a volume to the guest OS, it is possible that the event trigger for udev fails. In these cases, there is no additional trigger to ever rerun the scan. This can result in a failure indicated by `couldn't find device by serial id` and is unrecoverable without manual intervention.

This change address this by adding a udevadm trigger on a specified block device in the `getDeviceBySerialID` loop when it sees a device with an empty serial ID.

Fixes #152 

The unit testing added effectively creates a script called `udevadm` and add it to $PATH. This allows the unit testing to preempt the udevadm call within code and validate both input values and mock output values.

**Release note**:

```release-note
While this change will not break existing setup, for it to be effective, host `/sys` must be mounted in the NodeService Pods or be otherwise accessible.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ LABEL maintainers="The KubeVirt Project <kubevirt-dev@googlegroups.com>" \
 
 ENTRYPOINT ["./kubevirt-csi-driver"]
 
-RUN dnf install -y e2fsprogs xfsprogs && dnf clean all
+RUN dnf install -y e2fsprogs xfsprogs systemd-udevd && dnf clean all
 
 ARG git_sha=NONE
 LABEL multi.GIT_SHA=${git_sha}
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/openshift/build-machinery-go v0.0.0-20240613134303-8359781da660
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/net v0.33.0
+	golang.org/x/sys v0.28.0
 	google.golang.org/grpc v1.65.0
 	google.golang.org/protobuf v1.34.2
 	gopkg.in/yaml.v2 v2.4.0
@@ -63,7 +64,6 @@ require (
 	github.com/sirupsen/logrus v1.9.3 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
 	golang.org/x/oauth2 v0.21.0 // indirect
-	golang.org/x/sys v0.28.0 // indirect
 	golang.org/x/term v0.27.0 // indirect
 	golang.org/x/text v0.21.0 // indirect
 	golang.org/x/time v0.3.0 // indirect
--- a/pkg/service/node.go
+++ b/pkg/service/node.go
@@ -555,12 +555,62 @@ func getDeviceBySerialID(serialID string, deviceLister DeviceLister) (device, er
 	}
 
 	for _, d := range devices.BlockDevices {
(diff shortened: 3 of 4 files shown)


awels:
looks like we are missing some library inside either the infra OS or guest OS, can't tell from the error message

awels:
merged with https://github.com/kubevirt/csi-driver/pull/363

Issue #137, opened 2025-04-17
Is there a way to use this CSI without CAPK?


kreeuwijk:
**Is your feature request related to a problem? Please describe**:
The Kubevirt CSI driver requires the `.spec.providerID` to be populated on the node, which is only done by the Cluster API provider for Kubevirt (CAPK). That project isn't very mature yet and doesn't see frequent updates.

**Describe the solution you'd like**:
Allow the use of the CSI driver without the CAPK being present.

**Describe alternatives you've considered**:
None.

**Additional context**:
Our platform provides a non-CAPK way to deploy virtual K8S clusters on top of Kubevirt and we would like to add support for this CSI. But that's not possible as long as CAPK is required.

kreeuwijk:
Since the current solution is namespace-based, you could just go off the node hostname, assuming the VM is using the same name. That can be guaranteed in our solution at least. And since VMs must have unique names within a namespace, there's also no risk of name collisions.

awels:
@kreeuwijk Are you using CAPI to provision your clusters or do you have a completely different mechanism? It looks like CAPI is responsible for setting the providerID. If not, looking at some information about the node `.spec.providerID` it is entirely possible to add the providerID after the node is specified. So that might be an option as well.

Alex had a good reason for going with the providerID instead of the name. I just don't recall the details.

kreeuwijk:
We are not using CAPI to provision the virtual clusters on top of Kubevirt, but the bare metal cluster that hosts Kubevirt is using CAPI. We don't use CAPK yet since we have some concerns about the maturity of this provider. It doesn't properly support clusters from VMs that are running on Multus bridge networks for example.

We may build support for all of this later, but right now we have a non-CAPI solution for K8S clusters on Kubevirt that works well.

We could manually populate the providerID field in the VM node spec, but I would need to know what it should contain and if this would still work if the host cluster didn't contain the CAPK provider framework.

awels:
Okay looking at the code the `spec.ProviderID` should be of the format `kubevirt://<vm name>`. The code just [strips off the prefix and saves the vmName as part of the node ID](https://github.com/kubevirt/csi-driver/blob/main/cmd/kubevirt-csi-driver/kubevirt-csi-driver.go#L109-L124). Then later in the driver when we are determining if the volumes have been detached we use that information to find the correct VM.

kreeuwijk:
[...]

PR #170, opened 2025-12-31
Add annotation-based fallback for infra VM node ID resolution

orenc1:
Previously, the CSI driver required nodes to have a `providerID` prefixed
with `kubevirt://` to resolve the infra cluster VM mapping. This worked
for clusters provisioned by Cluster API with the KubeVirt provider, but
failed for other deployment scenarios where nodes have different
providerID formats (e.g., baremetalhost:///, aws:///, or empty).

This change introduces a fallback mechanism using node annotations when
the kubevirt:// providerID is not available:

  - csi.kubevirt.io/infra-vm-name: the VM name in the infra cluster
  - csi.kubevirt.io/infra-vm-namespace: the VM namespace in the infra cluster

Resolution order:
1. If providerID starts with "kubevirt://", extract VM name from it and
   namespace from the "cluster.x-k8s.io/cluster-namespace" annotation
2. Otherwise, use the csi.kubevirt.io annotations for both name and
   namespace

When resolution fails, the CSI node pod fails at registration with a
descriptive error message guiding users to configure the required
annotations and restart the pod. This ensures clear feedback about
misconfiguration rather than silent failures during volume operations.

Fixes # https://github.com/kubevirt/csi-driver/issues/137

**Release note**:

```release-note
Add annotation-based fallback for infra VM node ID resolution
```

### DIFF
--- a/cmd/kubevirt-csi-driver/kubevirt-csi-driver.go
+++ b/cmd/kubevirt-csi-driver/kubevirt-csi-driver.go
@@ -111,15 +111,11 @@ func handle() {
 		if err != nil {
 			klog.Fatal(fmt.Errorf("failed to find node by name %v: %v", nodeName, err))
 		}
-		if node.Spec.ProviderID == "" {
-			klog.Fatal("provider name missing from node, something's not right")
+
+		nodeID, err = resolveNodeID(*nodeName, node.Spec.ProviderID, node.Annotations)
+		if err != nil {
+			klog.Fatal(err)
 		}
-		vmName := strings.TrimPrefix(node.Spec.ProviderID, `kubevirt://`)
-		vmNamespace, ok := node.Annotations["cluster.x-k8s.io/cluster-namespace"]
-		if !ok {
-			klog.Fatal("cannot infer infra vm namespace")
-		}
-		nodeID = fmt.Sprintf("%s/%s", vmNamespace, vmName)
 		klog.Infof("Node name: %v, Node ID: %s", *nodeName, nodeID)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/service/node.go
```diff
@@ -506,7 +516,14 @@ func (n *NodeService) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
 	// the nodeID is the VM's ID in kubevirt or VMI.spec.domain.firmware.uuid
-	return &csi.NodeGetInfoResponse{NodeId: n.nodeID}, nil
+	// If node ID resolution failed, return a placeholder to allow CSI registration.
+	// Volume operations will fail with a descriptive error in ControllerPublishVolume.
+	nodeID := n.nodeID
+	if nodeID == "" && n.nodeIDErr != nil {
+		// Use format "unresolved/<nodeName>" so controller can include the node name in error messages
+		nodeID = "unresolved/" + n.nodeName
+	}
+	return &csi.NodeGetInfoResponse{NodeId: nodeID}, nil
```
akalenyu: I think we're over complicating things a bit; we can just exit on any errors to build a proper nodeID.
There is no point starting the services in that case IMO

orenc1: Good point.
The initial thought was to avoid the kubevirt-csi-node pods being in Crashloop due to this, which doesn't look nice, and the event on the pod that tries to mount the PVC will tell the users that the vm info could not be resolved.

I'll update the PR so the pods will fail instead.

cmd/kubevirt-csi-driver/kubevirt-csi-driver.go
```diff
@@ -184,3 +184,31 @@ func parseLabels() map[string]string {
 
 	return infraClusterLabelsMap
 }
+
+// resolveNodeID resolves the infra cluster VM name and namespace from the node's providerID or annotations.
+// It returns the nodeID in the format "namespace/name" or an error if resolution fails.
+func resolveNodeID(nodeName, providerID string, annotations map[string]string) (string, error) {
```
akalenyu: I think we can cut down even more, just exit (klog.Fatal I believe) if this returns an err. No point in starting the services (if the motivation is end user debugging, I doubt they need it)

orenc1: thanks. done

akalenyu (commented):
/approve
/lgtm

thank you!

PR #171, opened 2026-01-01
feat(storage): Support ReadWriteMany (RWX) filesystem volumes with pre-formatted backing

aagarwal-apexanalytix:

This PR adds support for `ReadWriteMany` (RWX) access mode with filesystem volumes (in addition to existing RWX raw block support).

Many users want shared filesystem storage across multiple VirtualMachines in KubeVirt, especially when backed by infra storage classes that support multi-attach/multi-writer (e.g., NFS-based or other shared filesystems). Previously, RWX filesystem was restricted to prevent unsafe `mkfs` races during concurrent multi-node staging.

This change enables safe RWX filesystem support by:
- Removing the provisioning-time rejection of blank RWX filesystem volumes.
- Modifying `NodeStageVolume` to skip filesystem creation for RWX volumes.
- Failing staging with a clear error if no filesystem is detected on RWX volumes (enforcing pre-formatting via infra StorageClass `fsType` parameter, or using a volume content source like clone/snapshot from a formatted volume).

This prevents corruption from concurrent `mkfs` while allowing RWX filesystem when the backing storage is properly initialized.

Non-RWX (RWO) filesystem volumes continue to auto-format blank devices as before. Raw block RWX remains unchanged and fully supported.

**Special notes**:
- Tested with infra StorageClasses that support RWX and pre-formatting (e.g., set `fsType: xfs` or `ext4` in the infra SC parameters).
- Verified blank RWX filesystem fails staging gracefully if not pre-formatted.
- Confirmed RWX filesystem works with clone/snapshot sources (pre-existing FS) and multi-node publish/mount.
- Raw block RWX and RWO filesystem behaviors remains unchanged.
- Consider adding e2e tests for RWX filesystem scenarios.

**Release note**:
```release-note
Add support for ReadWriteMany (RWX) filesystem volumes. Requires pre-formatted backing storage (via infra StorageClass fsType or volume content source) to ensure safe multi-node access.
```

### DIFF
--- a/pkg/service/controller.go
+++ b/pkg/service/controller.go
@@ -286,15 +286,22 @@ func (c *ControllerService) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	if err := c.validateDeleteVolumeRequest(req); err != nil {
 		return nil, err
 	}
+
 	dvName := req.VolumeId
 	klog.V(3).Infof("Removing data volume with %s", dvName)
 
 	err := c.virtClient.DeleteDataVolume(ctx, c.infraClusterNamespace, dvName)
 	if err != nil {
-		klog.Error("failed deleting DataVolume " + dvName)
+		if errors.IsNotFound(err) {
+			// Already deleted – idempotent success
+			klog.V(4).Infof("DataVolume %s already deleted (NotFound), treating as success", dvName)
+			return &csi.DeleteVolumeResponse{}, nil
+		}
+		klog.Errorf("failed deleting DataVolume %s: %v", dvName, err)
 		return nil, err
 	}
 
+	klog.V(4).Infof("Successfully deleted DataVolume %s", dvName)
 	return &csi.DeleteVolumeResponse{}, nil
 }
 
@@ -708,9 +715,22 @@ func (c *ControllerService) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 	if len(req.GetSnapshotId()) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "snapshot id missing in request")
 	}
-	if err := c.virtClient.DeleteVolumeSnapshot(ctx, c.infraClusterNamespace, req.GetSnapshotId()); err != nil {
+
+	snapshotName := req.GetSnapshotId()
+	klog.V(3).Infof("Deleting VolumeSnapshot %s", snapshotName)
+
+	err := c.virtClient.DeleteVolumeSnapshot(ctx, c.infraClusterNamespace, snapshotName)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Already deleted – idempotent success
+			klog.V(4).Infof("VolumeSnapshot %s already deleted (NotFound), treating as success", snapshotName)
+			return &csi.DeleteSnapshotResponse{}, nil
+		}
+		klog.Errorf("failed deleting VolumeSnapshot %s: %v", snapshotName, err)
 		return nil, err
 	}
+
+	klog.V(4).Infof("Successfully deleted VolumeSnapshot %s", snapshotName)
 	return &csi.DeleteSnapshotResponse{}, nil
 }
 


awels:
> > Something strange is going on. What happens on the KubeVirt side when the infra cluster is filesystem (RWX or RWO) is that a disk image file is created on the filesystem. That disk image file is then hotplugged into the guest as a block device. This should happen regardless of the underlying provider (CephFS/NFS/whatever). So you end up with infra (filesystem) -> disk image file -> hotplugged into guest -> then the driver creates a filesystem on the guest block device.
> > I don't see how cephfs on the infra could possibly influence the filesystem in the guest. You are saying the block device that is visible inside the guest contains the cephfs filesystem. Are we hotplugging the block device instead of the cephfs containing the disk image?
> 
> thanks for the detailed explanation - you're spot on about the standard behavior: for filesystem-mode infra volumes (CephFS, NFS, etc.), the KubeVirt/CDI stack typically creates a disk.img file on the mounted filesystem, then hotplugs that file as a block device to the guest VM. The guest then sees a raw block device and must format + mount it itself. The manifests I shared (and what I'm seeing in my test cluster) show a different - and actually very nice - outcome with this change/PR:

I would hope I am correct I wrote the hotplug code and most of the csi driver code.

> 
> The workload-cluster PVC (headlamp) uses volumeMode: Filesystem, accessModes: ReadWriteMany, provisioned via csi.kubevirt.io. It references storageClassName: nfs-csi, which has parameters: infraStorageClassName: ceph-filesystem-repl-2 (a CephFS RWX class). A prime PVC is created in the infra cluster using that CephFS StorageClass.

This should create a blank disk.img on the filesystem. We call `qemu-img create` to create the disk.img file. But it is definitely blank when we create it. Now I know cephfs does some funny business with sparse disk image files (for instance it reports the volume fully in use even when it is not). I might need to go ask the ceph folks about this.

> In the guest VM, after hotplug (as block device via SCSI), the device appears as a pre-formatted ext4 filesystem - ready to mount without any guest-side mkfs or manual formatting.
> 

This also sounds wrong, ext4 does not allow for concurrent access the meta data will get messed up when you have concurrent writes to an ext4 filesystem. Could you do me a favor? Can you create 2 pods in the guest cluster that use the same RWX filesystem volume, and just write some random data to both at the same time 

> Why this works so well: The PR/change modifies the provisioning/hotplug flow to bypass the disk.img layer when the infra StorageClass is a filesystem type (especially RWX ones like CephFS). Instead:
> 
> The Ceph CSI provisions a subvolume (or equivalent). The volume is formatted (ext4 in this case) during provisioning/population. It's hotplugged directly as a block device, but one that already contains a valid filesystem.

This is the part that is confusing me. The disk.img should be blank, we literally create the blank disk.img file in the prime pod. But you are telling me there is a filesystem on it. It could just be a weird interaction with cephfs and sparse disk images. I suspect that if you try this with a different provisioner on the infra cluster (lets say NFS) it will not work because the disk.img file will actually be blank.

> 
> This gives a much cleaner and more convenient end-to-end experience for RWX filesystem workloads - no extra formatting step inside the VM is required, similar to how NFS RWX volumes often behave today when presented via block hotplug in certain setups. It's particularly valuable when the infra storage already handles shared access reliably (as CephFS does here). Why this is a great fit for the current hotplug limitations: While we await virtio-fs enhancements for true filesystem passthrough, this approach leverages the existing block hotplug mechanism in a smart way: the storage backend does the heavy lifting (provisioning + formatting), and the guest gets a ready-to-use filesystem on a block device. It's a pragmatic optimization that delivers excellent UX without waiting for virtio-fs hotplug. In my testing, it's been stable, performant, and extremely convenient for shared filesystem use cases (e.g., the Headlamp app mounts it cleanly right away). I'm happy to share guest-side outputs (lsblk, blkid, mount, dmesg logs) or a minimal repro if it helps confirm the behavior. What are your thoughts - does this align with the intended direction of the PR? Ready to merge, or is there anything else you'd like to see before we proceed?
> 

We need to do a bit more testing here, in particular with different infra providers and making sure that if we do concurrent writes in pods in the guest cluster it also works since ext4 definitely is not a clustered filesystem.

> Sample yaml given below which works for me.
> 
> Thanks again for the great discussion!
>

awels:
Trust me, we would very much like to add support for RWX filesystem volumes to this csi-driver. We have spend quite a bit of time trying to figure out how to do it effectively. But it all boils down to that we have no good way of passing through filesystems from the infra cluster. Or that we don't have a clustered filesystem available to put on the block devices we make available. But you did give me an interesting idea with this, if we can directly pass through the block device that is underlying the filesystem to the guest, then the filesystem would be visible inside the guest. Something to investigate for sure. There could be some issues if the infra node is mounting the same filesystem already, but we could potentially unmount on the infra node.

PR #180, opened 2026-04-24
Fix silent error swallowing in makeFS and getDeviceBySerialID

carterpewpew:

Both `makeFS` and `getDeviceBySerialID` in `pkg/service/node.go` performed a type assertion to `*exec.ExitError` before checking if `err` was nil. When the error was non-nil but not an `*exec.ExitError` (e.g., `exec.ErrNotFound` when `mkfs` or `lsblk` is missing from `PATH`), the error-handling block was skipped entirely:
- **`makeFS`**: returned `nil`, causing `NodeStageVolume` to report success and mark the volume as staged **without a filesystem**. The volume became permanently unusable because kubelet would not retry a successful Stage, and subsequent `NodePublishVolume` mount calls would fail with a confusing error.
- **`getDeviceBySerialID`**: fell through to `json.Unmarshal` on nil output, replacing the real error with a misleading JSON parse error and misdirecting debugging.
This PR:
- Reorders error handling to check `err != nil` first, then attempts the `*exec.ExitError` match inside the guarded block
- Uses `errors.As` instead of direct type assertion for robustness against wrapped errors
- Fixes a pre-existing copy-paste bug that logged `stdout` in place of `stderr` in `makeFS`
- Fixes a missing space in the `getDeviceBySerialID` error message
- Adds unit tests for non-ExitError propagation in both functions
- Adds an ExitError regression test to ensure enriched error messages (with stderr details) still work

Fixes #

No behavioral change for the happy path or the ExitError path. The only change is that errors which were previously swallowed now correctly propagate.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/service/node.go
+++ b/pkg/service/node.go
@@ -542,9 +542,12 @@ func getDeviceBySerialID(serialID string, deviceLister DeviceLister) (device, er
 	klog.Infof("Get the device details by serialID %s", serialID)
 
 	out, err := deviceLister.List()
-	exitError, incompleteCmd := err.(*exec.ExitError)
-	if err != nil && incompleteCmd {
-		return device{}, errors.New(err.Error() + "lsblk failed with " + string(exitError.Stderr))
+	if err != nil {
+		var exitError *exec.ExitError
+		if errors.As(err, &exitError) {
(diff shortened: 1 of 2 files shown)


awels (commented):
/test all
/lgtm
/approve

PR #184, opened 2026-06-04
Check VMI status when detaching, not just VM spec

### LINKED ISSUE #182: ControllerUnpublishVolume leaks orphan hot-plug volumes when VM teardown precedes VMI termination

/kind bug

**What happened**:

In production we observe hot-plug "zombie" pods that survive after a tenant Pod is rescheduled from one tenant Node (KubeVirt VM A) to another (VM B). The CSI driver issues `RemoveVolume` to VM A and `AddVolume` to VM B, but in several edge cases the VM-A detach is silently dropped. The hot-plug pod on VM A stays Running, the underlying infra storage device stays exclusively attached to the source host, and the new attach on VM B fails with:

```
MapVolume.MapPodDevice failed for volume "pvc-XXXX" : rpc error: code = Internal
desc = NodePublishVolume failed for pvc-XXXX: failed to set source device readwrite: exit status 1
```

(on DRBD-backed infra storage; other RWO backends surface this as `Multi-Attach error for volume "pvc-XXXX"` from the kube-controller-manager attach/detach controller — same root cause).

We hit this 12 times concurrently across two tenant clusters today; the manual workaround is `virtctl removevolume <vm-A> -n <ns> --volume-name=<dvName>` for each, after which the new attach succeeds.

`ControllerUnpublishVolume` should always release the hot-plug from the VMI, even when the parent VM object is missing or out of sync with the VMI status. Returning success without doing so leaves orphan hot-plug pods that are unsafe for any storage backend enforcing exclusive attachment.

1. Tenant cluster running on top of KubeVirt (e.g. via CAPK).
2. Deploy a StatefulSet with one RWO PVC backed by the kubevirt-csi driver (`storageClassName: kubevirt`).
3. Cordon + drain the tenant Node hosting the pod so the StatefulSet pod moves to a different tenant Node (i.e. a different KubeVirt VM in the infra cluster).
4. While step 3 is in flight, force-delete the source `VirtualMachine` object (`kubectl delete vm <vm-A>`) **before** its VMI finishes terminating. This simulates CAPI-driven `KubeVirtMachine` deletion.
5. Observe:
   - The new tenant Pod stays in `ContainerCreating`.
   - `kubectl describe pod hp-volume-XXXX -n <ns>` on the destination reports `MapVolume.MapPodDevice failed ... failed to set source device readwrite`.
   - The source VMI (still alive) keeps the hot-plug in `.status.volumeStatus[].hotplugVolume`; the hot-plug pod on that VMI is still Running.

It is also reproducible without the force-delete, just less reliably: virt-api updates `VM.spec` optimistically on `removevolume` but virt-handler can take seconds to actually unplug; if external-attacher gives up in that window the detach is never re-attempted.

### Root cause

`pkg/service/controller.go` `ControllerUnpublishVolume`:

```go
_, err = c.virtClient.GetWorkloadManagingVirtualMachine(ctx, c.infraClusterNamespace, vmName)
if err != nil {
    if !errors.IsNotFound(err) {
        return nil, err
    }
    // VM removed, vacuously succeeded in unpublish
    return nil, nil
}
```

That short-circuit was introduced by #119 to fix #83 ("volume attached to a node that was removed could not be detached"). It correctly unblocks the cleanup of a `VolumeAttachment` whose Node is permanently gone, but when the VMI (and its hot-plug pod, and the infra storage attachment) outlive the VM by even a few seconds — which happens routinely during CAPK machine deletion, PVC finalizer waits, or live-migration — the synchronous detach path silently skips work that still needs to be done.

The same kind of false-positive lives inside `EnsureVolumeRemovedVM` (`pkg/kubevirt/client.go`) which returns `(true, nil)` on VM-not-found, and inside `removeVolumeFromVm` (`pkg/service/controller.go`) which exits early before reaching the existing VMI-fallback branch added by #134.

### Proposed fix

Make `ControllerUnpublishVolume` and `removeVolumeFromVm` consult **both** `VM.spec.template.spec.volumes` and `VMI.status.volumeStatus`, short-circuiting only when both surfaces agree the volume is gone. A new `EnsureVolumeRemovedVMI` helper would mirror the existing `EnsureVolumeRemovedVM` API.

I have a PR ready for this.

### Environment

- KubeVirt CSI driver built from a recent rebase of upstream `main` (verified against commit `51c6bca`).
- KubeVirt: v1.8.x in the infra cluster, CAPK-managed VMs.
- Infra storage: LINSTOR + DRBD 9.2.x. The same code path also breaks on any infra StorageClass that enforces single-writer semantics.

Happy to provide raw logs, VMI/VM YAML dumps, or to test patches.

mattia-eleuteri:

`ControllerUnpublishVolume` currently returns success as soon as the parent `VirtualMachine` object is missing, on the assumption that a missing VM means the hot-plug is already gone. In practice the VMI (and its hot-plug pod) can outlive the parent VM during a CAPI/CAPK teardown race — the infra-side storage device then stays exclusively attached to the source host, and the next attempt to attach the same volume to a different VM fails with `Multi-Attach error` or, on backends like DRBD that enforce exclusive attachment, with `failed to set source device readwrite`.

The same divergence appears when virt-api updates `VM.spec` optimistically on a `removevolume` call but virt-handler has not yet unplugged the QEMU device.

This PR:

- Adds `EnsureVolumeRemovedVMI` to the `kubevirt.Client` interface, mirroring the existing `EnsureVolumeRemovedVM` but inspecting `VMI.status.volumeStatus` instead of `VM.spec.template.spec.volumes`.
- Replaces the `VM not found = success` short-circuit in `ControllerUnpublishVolume` with a new `volumeStillAttached` helper that consults both surfaces. The original #119 unblock is preserved (the call still succeeds when the volume is genuinely gone from both); the new behaviour is that we now actually detach from the VMI when it still holds the hot-plug.
- Refactors `removeVolumeFromVm` so that it falls through to `RemoveVolumeFromVMI` when the parent VM is gone but the VMI still reports the hot-plug, closing the orphan path end-to-end.
- Adds a regression test in `controller_test.go` that covers the "VM gone but VMI still has the hot-plug" path.

We hit the bug 12 times across two tenant clusters in a single day before opening the linked issue. Full failure-mode catalogue and reproducer are in the issue.

Fixes #182
Refs #83 #119 #134

- This changes behaviour: a "VM not found" no longer means "detach succeeded". For existing deployments that have already lost the VM object and only have the orphan VMI left, `ControllerUnpublishVolume` will now actually run `RemoveVolumeFromVMI` against the VMI — i.e. it will do the right thing instead of silently doing nothing. There is no CSI/API-surface change.
- I added one method (`EnsureVolumeRemovedVMI`) to the `kubevirt.Client` interface. I did **not** update `pkg/kubevirt/mocked_client.go`: the file is already stale relative to the interface (missing `EnsureVolumeRemovedVM` and `EnsureVolumeAvailableVM` among others) and `make mockgen` produces a much larger diff than the orphan-volume fix justifies. A separate "regen mocks" PR would be the right home for that. The hand-written test fakes (`ControllerClientMock` in `pkg/service/controller_test.go` and `fakeKubeVirtClient` in `sanity/sanity_test.go`) are updated.
- A companion PR (#185, separate and independent) introduces an opt-in `DetachReconciler` to catch orphans this synchronous path cannot reach (external-attacher giving up after force-detach timeout, VMIs in mid-migration, legacy state accumulated before this fix). They can be reviewed and merged in any order.

**Release note**:
```release-note
BugFix: ControllerUnpublishVolume no longer treats a missing parent VirtualMachine as "detach succeeded" — it now also checks the VMI status and detaches the hot-plug from the VMI if one is still attached, preventing orphan hot-plug pods that pin the infra storage device exclusively to the source host.
```

### DIFF
--- a/pkg/kubevirt/client.go
+++ b/pkg/kubevirt/client.go
@@ -70,6 +70,7 @@ type Client interface {
 	EnsureVolumeAvailableVM(ctx context.Context, namespace, name, volumeName string) (bool, error)
 	EnsureVolumeRemoved(ctx context.Context, namespace, vmName, volumeName string, timeout time.Duration) error
 	EnsureVolumeRemovedVM(ctx context.Context, namespace, name, volumeName string) (bool, error)
+	EnsureVolumeRemovedVMI(ctx context.Context, namespace, name, volumeName string) (bool, error)
 	EnsureSnapshotReady(ctx context.Context, namespace, name string, timeout time.Duration) error
 	EnsureControllerResize(ctx context.Context, namespace, claimName string, timeout time.Duration) error
 	CreateVolumeSnapshot(ctx context.Context, namespace, name, claimName, snapshotClassName string) (*snapshotv1.VolumeSnapshot, error)
@@ -598,6 +599,34 @@ func (c *client) EnsureVolumeRemovedVM(ctx context.Context, namespace, name, vol
 			continue
 		}
 		if volume.Name == volumeName {
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+// EnsureVolumeRemovedVMI returns true when the VMI no longer reports the
+// volume as a hot-plug in its status. It is the live-state counterpart of
+// EnsureVolumeRemovedVM, which only inspects the VM spec. The two diverge
+// when the parent VM has been deleted but the VMI (and its hot-plug pod)
+// is still around, or when virt-api mutated VM.spec optimistically but
+// virt-handler has not unplugged the device yet.
+func (c *client) EnsureVolumeRemovedVMI(ctx context.Context, namespace, name, volumeName string) (bool, error) {
+	vmi, err := c.virtClient.KubevirtV1().VirtualMachineInstances(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return false, err
+		}
+		// No VMI, vacuously removed
+		return true, nil
+	}
+
+	for _, volumeStatus := range vmi.Status.VolumeStatus {
+		if volumeStatus.HotplugVolume == nil {
+			continue
+		}
+		if volumeStatus.Name == volumeName {
 			return false, nil
 		}
 	}
--- a/pkg/service/controller.go
+++ b/pkg/service/controller.go
@@ -534,22 +534,17 @@ func (c *ControllerService) ControllerUnpublishVolume(ctx context.Context, req *
 		klog.Error("failed getting VM Name for node ID " + req.NodeId)
 		return nil, err
 	}
-	_, err = c.virtClient.GetWorkloadManagingVirtualMachine(ctx, c.infraClusterNamespace, vmName)
-	if err != nil {
-		if !errors.IsNotFound(err) {
-			return nil, err
-		}
-		// VM removed, vacuously succeeded in unpublish
-		return nil, nil
-	}
-
-	// Fast-path: if the disk is no longer on the VM, succeed immediately
-	notPresent, err := c.virtClient.EnsureVolumeRemovedVM(ctx, c.infraClusterNamespace, vmName, dvName)
+	// We do NOT short-circuit on "VM not found" anymore. The VMI (and its
+	// hot-plug pod) can outlive the parent VM during a CAPI/CAPK teardown
+	// race; returning success here used to leave orphan hot-plug pods
+	// that hold the infra storage device exclusively attached to the source
+	// host — blocking subsequent attachments. See kubevirt/csi-driver#83.
+	attached, err := c.volumeStillAttached(ctx, vmName, dvName)
 	if err != nil {
 		return nil, err
 	}
-	if notPresent {
-		klog.V(3).Infof("Volume %s already detached from VM %s – skipping hot-unplug", dvName, vmName)
+	if !attached {
+		klog.V(3).Infof("Volume %s already detached from VM/VMI %s – skipping hot-unplug", dvName, vmName)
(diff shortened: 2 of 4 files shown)


mattia-eleuteri:
Reopening to retrigger DCO check (the check didn't fire on initial open, likely a webhook race).

kvaps (approved):
The diagnosis is correct: `ControllerUnpublishVolume` currently short-circuits to success as soon as the parent VM is gone (and even returns `nil, nil` there), so an orphan VMI keeps its hot-plug pod alive and the infra device stays attached to the source host. Checking both surfaces (`VM.spec` via `EnsureVolumeRemovedVM` and `VMI.status.volumeStatus` via the new `EnsureVolumeRemovedVMI`) and falling through to `RemoveVolumeFromVMI` when the VM is missing is the right fix, and the final `EnsureVolumeRemoved` wait was already VMI-aware, so the detach semantics stay consistent end-to-end.

One behavioural note, not a blocker: during a normal unplug, after the volume is gone from `VM.spec` but before virt-handler finishes the unplug, a retried unpublish now goes down the VMI branch and calls the `removevolume` subresource against a volume that is already mid-unplug. The subresource will likely reject that, the backoff loop swallows the error and the external-attacher retries, so it converges — just expect some extra log noise compared to the old (incorrectly optimistic) fast path.

Not regenerating `mocked_client.go` seems reasonable given the file is already stale relative to the interface; a separate mock-regen PR is the right place for that.

LGTM overall — we hit this exact failure mode with DRBD-backed storage, where the leaked hot-plug pod keeps the device exclusively attached and every subsequent attach fails with `failed to set source device readwrite`.

awels (approved):
/approve

PR #379, opened 2026-08-21
Allow RWX Filesystem volumes via allowRWXFilesystem StorageClass parm

m-aamir-ashfaq:
### What this PR does / why we need it:
RWX Filesystem volumes (ReadWriteMany + volumeMode: Filesystem) are currently
hard-blocked in validateCreateVolumeRequest regardless of the infra StorageClass.

This guard is correct for RBD — two nodes mounting the same RBD image as a
filesystem simultaneously causes data corruption. It is too broad for CephFS,
which natively supports multi-writer filesystem mounts.

This PR does not remove the guard. It adds an explicit opt-in StorageClass
parameter: allowRWXFilesystem=true. The default behaviour is unchanged — RWX
Filesystem is blocked unless the admin explicitly sets this parameter. No
existing user is affected.

### Boundary conditions:
- Without the flag: identical to today. No regression.
- With the flag on CephFS-backed infra StorageClass: works correctly.
  ControllerPublishVolume already handles multi-node attach for RWX — the isRWX
  check skips the "already attached to another VMI" guard, so the same DataVolume
  can be hotplugged into multiple worker VMs independently.
- With the flag on RBD-backed infra StorageClass (misconfiguration): the infra
  PVC fails to bind — RBD does not advertise RWX support. Clean failure, no data
  corruption.
- An admission policy (e.g. Gatekeeper) can be used to enforce that
  allowRWXFilesystem=true is only set on StorageClasses backed by a
  filesystem-capable infra StorageClass such as CephFS.
**Which issue(s) this PR fixes:**
Fixes #
(no upstream issue exists yet)
**Special notes for your reviewer:**
PR #107 (acc2730) fixed RWX Block — when a guest requests RWX, the infra
DataVolume is created with RWX AccessModes, enabling live migration. It did not
modify the RWX Filesystem guard. The guard is present and unchanged in all
versions including current main. This PR addresses the remaining gap.
```release-note
Allow RWX Filesystem volumes by setting allowRWXFilesystem=true in StorageClass
parameters. Default behaviour is unchanged. Only use with filesystem-capable
infra StorageClasses such as CephFS.
```


m-aamir-ashfaq:
Closing this PR — fundamental architectural issue discovered after submission.

PR #382, opened 2026-09-22
Add flag to disable the VMI hot-unplug fallback

### LINKED ISSUE #381: Volume detach falls back to the VMI subresource, which needs the deprecated HotplugVolumes gate
`removeVolumeFromVm` takes the volume out of the VM object, and when the volume is not in `VM.spec` it falls back to `RemoveVolumeFromVMI`. That fallback is the path a control-plane teardown takes, and it only works while KubeVirt's `HotplugVolumes` feature gate is on.

KubeVirt marked `HotplugVolumes` deprecated in v1.9.0 (`pkg/virt-config/featuregate/inactive.go`), replaced by `DeclarativeHotplugVolumes`. The two are mutually exclusive at runtime: `handleDeclarativeVolumeHotplug` skips the declarative path entirely while `HotplugVolumes` is enabled, so a cluster cannot move to the replacement without turning the old gate off, and turning it off is what breaks this fallback. `virt-api` refuses a VMI-level `removevolume` on a VM-owned VMI unless `ephemeralHotplugSupported()` is true, and that function returns exactly `HotplugVolumesEnabled()`.

The result for a downstream shipping both: it cannot enable the replacement gate until this driver stops needing the VMI subresource, and it cannot drop the deprecated one either. Cozystack is in that position today (cozystack/cozystack#4032).

I have read #112 and the VM-level path that came with #184, which is what we pin. This is about what is left underneath it: the VMI fallback and the gate it still implies.

What I am after is a direction rather than a specific patch. Whether the fallback can be dropped, or made conditional on the gate, or replaced with something that works when only `DeclarativeHotplugVolumes` is on. I can test a change against a tenant-cluster teardown, which is where the fallback actually fires for us.

Versions: KubeVirt v1.9.0, csi-driver pinned at v0.0.0-20260624122948-27b52aa22da7.

lexfrei:

This adds a controller flag that turns off the VMI-level hot-unplug fallback, so a cluster can move from the old hotplug feature gate to the declarative one. When a volume is no longer in `VM.spec`, `removeVolumeFromVm` calls `RemoveVolumeFromVMI`. virt-api accepts that call on a VM-owned VMI only while the `HotplugVolumes` gate is on. The check is `vmiVolumePatch` in `pkg/virt-api/rest/volumes.go`, which returns BadRequest unless `ephemeralHotplugSupported()`, and that is just `HotplugVolumesEnabled()`.

The flag is `--enable-vmi-hotplug-fallback`. It defaults to `true`, so an upgrade changes nothing.

With the flag set to `false`, the driver still removes the volume from `VM.spec` when it is there. If it is not there, or the VM is gone, the driver skips the VMI call for a VM-owned VMI and treats the volume as unplugged. `ControllerUnpublishVolume` still waits in `EnsureVolumeRemoved` until the VMI status drops the volume, so the detach is not reported early.

KubeVirt handles the rest on its side:

- With `DeclarativeHotplugVolumes`, virt-controller reconciles the VMI hotplug volumes against `VM.spec` (`handleDeclarativeVolumeHotplug` in `pkg/virt-controller/watch/vm/vm.go`). This also covers volumes left over from the old VMI-only hotplug while the VM exists.
- The VMI has a controller ownerReference to its VM, set in `SetupVMIFromVM`. Once the VM is deleted, the garbage collector removes the VMI under the default propagation.

Fixes #381

- I'm fine with renaming the flag. I picked this name because it says what `true` does.
- I didn't touch `deploy/` or `pkg/generated/bindata.go`. The default keeps current behaviour, and setting the flag there would mean regenerating bindata. The README gets one paragraph under "Configuring KubeVirt".
- A VM deleted with `propagationPolicy=Orphan` leaves a VMI with no owner, and nothing in KubeVirt reconciles its volumes. That VMI keeps the fallback even with the flag off, since virt-api accepts the call for it. The ownership check is a copy of `ownedByVirtualMachine` from virt-api.
- Unit tests cover flag on, flag off and the default, and sanity passes. I haven't run e2e with `DeclarativeHotplugVolumes` enabled.

**Release note**:

```release-note
Add the --enable-vmi-hotplug-fallback controller flag. Set it to false on clusters that use the KubeVirt DeclarativeHotplugVolumes feature gate instead of HotplugVolumes. The default (true) keeps the current behaviour.
```


lexfrei:
@awels this is the flag from #381. Prow does not let me trigger the tests, could you run `/test all`?

lexfrei:
The two failures are not from this change. Both jobs stop before running anything: the image cannot install the Go version from `toolchain go1.24.13` in go.mod ("I don't have any idea what to do with '1.24.13'", then `go: command not found`). The linter and split-e2e runs on #380 fail the same way, so main is affected too. The unit and sanity jobs passed here, and the other two e2e jobs are still running.

awels:
/retest-required
lets try again, if it fails the same way again, will have to figure out what changed in PROW.

lexfrei:
@awels thanks, the linter went green on the retest and split-e2e is still running. I can't trigger a rerun myself, so if split-e2e goes red again on the Go download, one more `/retest-required` should do it. For the download itself I opened kubevirt/project-infra#5488: it retries gimme a few times and stops the job at setup with a clear message instead of running on without Go.

lexfrei:
@awels split-e2e got Go this time and ran, and 17 of 18 specs passed. The one failure is the Block multi-attach spec: the second pod's volume attached, but `MapPodDevice` kept failing with "couldn't find device by serial id" until the 180s timeout. That is the attach path, which looks like #152. This PR only changes the unplug fallback in `removeVolumeFromVm`, and the default keeps today's behaviour. Could you give it one more `/retest-required`?

awels:
Great all passed, let me do a review.

awels (commented):
/approve

awels (approved):

